PR #1, opened 2022-02-11
Adopt new SwiftPM plugin Argument Parsing API

ethan-kusters:
## Summary

Updates the SwiftPM plugin to use the latest Argument Parsing API that will ship with Swift 5.6.

This also includes some general improvements to help output (`--help`) in preparation for the 1.0 release of the plugin.

## Details

The SwiftPM plugin API has been updated so that options like `--target` and `--product` are passed directly to the plugin instead of to the package manager.

This means that instead of calling the plugin with something like:

    swift package --target SwiftMarkdown generate-documentation

you would now do:

    swift package generate-documentation --target SwiftMarkdown

This allows us to be a little smarter about choosing which products to build. For example, now when a user invokes the basic

    swift package generate-documentation

we will also generate documentation for dependencies. Similarly, this allows clients to specify products of dependencies which makes doing something like

    swift package preview-documetnation --product ArgumentParser

possible for packages that _import_ ArgumentParser.

## Dependencies

This change depends on the latest Swift 5.6 nightly toolchain that includes the new ArgumentParsing API.

## Testing

Add the version of the plugin on this branch as a dependency of a package with:

```swift
.package(url: ["https://github.com/apple/swift-docc-plugin](https://github.com/ethan-kusters/swift-docc-plugin.git)", branch: "adopt-new-argument-parsing-api")
```

## Checklist

- [x] Added tests


Plugins/Swift-DocC Convert/SwiftDocCConvert.swift
```diff
@@ -58,17 +71,24 @@ import PackagePlugin
...
+            
             // Run `docc convert` with the generated arguments and wait until the process completes
             let process = try Process.run(doccExecutableURL, arguments: doccArguments)
             process.waitUntilExit()
             
+            let conversionDuration = conversionStartTime.distance(to: .now())
+            
             // Check whether the `docc convert` invocation was successful.
             if process.terminationReason == .exit && process.terminationStatus == 0 {
+                print("Conversion complete! (\(conversionDuration.descriptionInSeconds))")
```
franklinsch, line 85: I assume we're doing this because it aligns with other build phases of SwiftPM? I'm wondering if this should be the responsibility of SwiftPM for all plugins.

ethan-kusters: > I assume we're doing this because it aligns with other build phases of SwiftPM?

That's right- so now our output blends in a more understandable way, making it clear what's being emitted by DocC and what's being emitted by SwiftPM:

For example:

    > swift package generate-documentation
    Generating documentation for 'ArgumentParser'...
    Building for debugging...
    Build complete! (0.26s)
    Converting documentation...
    Conversion complete! (0.47s)
    Generated DocC archive at '/swift-markdown/.build/plugins/Swift-DocC/outputs/ArgumentParser.doccarchive'.

    Generating documentation for 'Markdown'...
    Building for debugging...
    Build complete! (0.28s)
    Converting documentation...
    /swift-markdown/Sources/Markdown/Markdown.docc/Markdown/BlockDirectives.md:149:146: warning: Topic reference 'Diagnostic' couldn't be resolved. No local documentation matches this reference.
    /swift-markdown/Sources/Markdown/Markdown.docc/Markdown/BlockDirectives.md:149:162: warning: Topic reference 'DiagnosticEngine' couldn't be resolved. No local documentation matches this reference.
    /swift-markdown/Sources/Markdown/Markdown.docc/Markdown/BlockDirectives.md:149:188: warning: Topic reference 'DiagnosticConsumer' couldn't be resolved. No local documentation matches this reference.
    /swift-markdown/Sources/Markdown/Base/Document.swift:14:5: warning: Topic reference 'child(through:)-65k60' couldn't be resolved. No local documentation matches this reference.
    Conversion complete! (0.95s)
    Generated DocC archive at '/swift-markdown/.build/plugins/Swift-DocC/outputs/Markdown.doccarchive'.

>  I'm wondering if this should be the responsibility of SwiftPM for all plugins.

I agree this would be a nice enhancement for a future release! Would just need an API to call to tell SwiftPM when starting and stopping build-like operations. CC: @abertelrud 

I don't think we should block implementing it in DocC in the meantime though.

Plugins/Swift-DocC Convert/SwiftDocCConvert.swift
```diff
@@ -35,7 +41,14 @@ import PackagePlugin
...
         
         // Iterate over the Swift source module targets we were given.
-        try targets.lazy.compactMap(\.asSwiftSourceModuleTarget).forEach { target in
+        for (index, target) in swiftSourceModuleTargets.enumerated() {
+            if index != 0 {
+                // Emit a line break if this is not the first target being built.
+                print()
+            }
+            
+            print("Generating documentation for '\(target.name)'...")
```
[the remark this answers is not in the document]
ethan-kusters: Ah that's fair but SwiftPM uses three separate periods so it looks odd if we don't:

```
Generating documentation for 'markdown-tool'…
Building for debugging...
Build complete! (0.29s)
Converting documentation…
```

Sources/SwiftDocCPluginUtilities/HelpInformation.swift
```diff
@@ -72,24 +66,48 @@ public enum HelpInformation {
...
+        USAGE: swift package --disable-sandbox [<package-manager-option>] preview-documentation [<plugin-options>] [<docc-options>]
+        
         NOTE: This plugin is only able to preview a single target at a time. If your
               package contains more than one documentable target, you must specify which
-              target should be previewed with the -target option.
+              target should be previewed with the --target or --product option.
+        
+        PACKAGE MANAGER OPTIONS:
+          --disable-sandbox
+                                  Disable using the sandbox when executing subprocesses
```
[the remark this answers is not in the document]
ethan-kusters: Good catch- thanks! Fixed here: 4dfdcdc0221ebbc9b37fc75a2219982acbc78792.

> Also, formatting for the lines below look a bit off.
Hm... What do you mean?

Plugins/SharedPackagePluginExtensions/DocCCommandPlugin.swift
```diff
@@ -0,0 +1,37 @@
...
+protocol DocCCommandPlugin: CommandPlugin {
+    func performDocCCommand(context: PluginContext, arguments: [String]) throws
+}
+
+extension DocCCommandPlugin {
+    func performCommand(context: PluginContext, arguments: [String]) throws {
+        do {
+            // We're wrapping this in a do/catch block instead of throwing directly
+            // from this function because SwiftPM doesn't currently print the localized
+            // error description of a thrown error.
```
tomerd, line 20: @abertelrud this sound like a bug to me. since the main method is throwing SwiftPM should catch that and emit the diagnostics w/o the plugin needing to wrap as such

cc @rauhul

Plugins/Swift-DocC Preview/SwiftDocCPreview.swift
```diff
@@ -115,8 +115,34 @@ import PackagePlugin
...
+        process.executableURL = doccExecutableURL
+        process.arguments = doccArguments
+        
+        // Monitor for a termination signal and pass any along to the child `docc` preview
+        // process.
+        //
+        // Since Foundation's process does *not* create new processes with the same group ID
+        // as the parent process, we don't expect this to happen automatically.
+        signal(SIGTERM, SIG_IGN)
+        let terminateSignalSource = DispatchSource.makeSignalSource(signal: SIGTERM)
```
ethan-kusters, line 129: @abertelrud I think ideally we would only need the `SIGINT` monitor to make this work but it looks like SwiftPM isn't giving plugins a chance to exist gracefully upon an interrupt signal before terminating them. Is that potentially a bug?

IntegrationTests/Tests/SingleLibraryTargetTests.swift
```diff
@@ -54,8 +52,8 @@ final class SingleLibraryTargetTests: XCTestCase {
         
         let result = try swiftPackage(
             """
-            --disable-sandbox generate-documentation \
-                --output-path "\(customOutputDirectory.path)"
+            --allow-writing-to-directory "\(customOutputDirectory.path)" \
```
abertelrud, line 55: Can this ever contain quotes that need to be escaped?  It's an edge case, to be sure, but figured I'd ask.

Plugins/SharedPackagePluginExtensions/DocCCommandPlugin.swift
abertelrud: Agreed, that seems like a bug.  I thought there was even a unit test in SwiftPM that checked that thrown errors got propagated correctly.

abertelrud: Yes, this test checks that: https://github.com/apple/swift-package-manager/blob/741994b58270cacc7efe319536d4df76dc86d966/Tests/CommandsTests/PackageToolTests.swift#L1350

But I wonder if `localizedDescription` is different somehow.  I think I've run into that before in SwiftPM itself, where stringifying the error just uses description but not localized description.  This seems worth a bug report though.

Plugins/Swift-DocC Convert/SwiftDocCConvert.swift
abertelrud: I agree — this actually came up in the discussion during the evolution proposal API review, in regard to providing progress.  I think we'd want to provide structured progress from the plugin, and that would include what conceptual tasks and subtasks it is carrying out.  SwiftPM should really use the same kind of machinery, and then all the output would be formatted the same way.  For now we don't have it and so it would be in a 2.0 of the plugin API.  I think DocC would be a great test case for it though, compared with some of the simpler plugins.

Plugins/SharedPackagePluginExtensions/DocCCommandPlugin.swift
ethan-kusters: I filed rdar://89327223 to track this and moved from `LocalizedError` to `CustomStringConvertible` in the mean time: d8976ddb.

IntegrationTests/Tests/Utility/XCTestCase+swiftPackage.swift
```diff
@@ -74,4 +83,8 @@ extension Pipe {
             String(data: $0, encoding: .utf8)
         }
     }
+    
+    var availableOutput: String? {
+        return String(data: fileHandleForReading.availableData, encoding: .utf8)
```
abertelrud, line 88: For testing purposes (as this is), this is probably fine.  But in general the available data might include only the beginning of a UTF-8 byte sequence, with the remaining bytes to follow later.  So to be accurate here you'd need to buffer any remaining partial sequence.  I don't remember the APIs offhand for that, but I think there is support in Swift stdlib for doing that.

Plugins/Swift-DocC Preview/SwiftDocCPreview.swift
abertelrud: Yes possibly.  It looks like SwiftPM is using `Foundation.Process.terminate()` — I think it should probably use `Foundation.Process.interrupt()`?

Sources/SwiftDocCPluginUtilities/FoundationExtensions/String+singleQuoted.swift
```diff
@@ -6,10 +6,10 @@
...
 
-import PackagePlugin
+import Foundation
 
-extension Target {
-    var asSwiftSourceModuleTarget: SwiftSourceModuleTarget? {
-        return self as? SwiftSourceModuleTarget
+extension String {
+    var singleQuoted: String {
+        return "'\(self)'"
```
abertelrud, line 13: Ideally this should probably escape any single quote and reverse solidus that's already in the string.

abertelrud (approved):
Looks good to me.  Thanks for trying out the SwiftPM plugin APIs!

shahmishal:
@swift-ci test

ethan-kusters:
@swift-ci please test

ethan-kusters:
@swift-ci please test

ethan-kusters:
@swift-ci please test

ethan-kusters:
@swift-ci please test

ethan-kusters:
@swift-ci please test

ethan-kusters:
@swift-ci please test

ethan-kusters:
@swift-ci please test

Sources/SwiftDocCPluginUtilities/FoundationExtensions/String+singleQuoted.swift
ethan-kusters: That's a good point. This matches the behavior of `docc` when emitting information like this so I think I'll leave it as-is for now but that would be a good future enhancement.

ethan-kusters:
@swift-ci please test

ethan-kusters:
@swift-ci please test

ethan-kusters:
@swift-ci please test

ethan-kusters:
@swift-ci please test

ethan-kusters:
@swift-ci please test

ethan-kusters:
@swift-ci please test

ethan-kusters:
@swift-ci please test

ethan-kusters:
@swift-ci please test

PR #7, opened 2022-03-14
Snippets

bitjammer:
## Summary

Add snippets support to the Swift DocC Plugin.

## Dependencies

No outstanding PR dependencies, although this does add a new dependency on SymbolKit, which should match DocC's version.

## Testing

New unit and integration tests added. Run tests as usual.

## Checklist

- [x] Added tests

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -66,9 +66,7 @@ let package = Package(
         .executableTarget(
             name: "snippet-build",
             dependencies: [
-                .product(name: "ArgumentParser", package: "swift-argument-parser"),
                 .product(name: "SymbolKit", package: "swift-docc-symbolkit"),
-                .product(name: "TSCBasic", package: "swift-tools-support-core"),
             ])
     ]
 )
@@ -76,13 +74,9 @@ let package = Package(
 if ProcessInfo.processInfo.environment["SWIFTCI_USE_LOCAL_DEPS"] == nil {
     package.dependencies += [
         .package(url: "https://github.com/apple/swift-docc-symbolkit", branch: "main"),
-        .package(url: "https://github.com/apple/swift-argument-parser", .upToNextMinor(from: "1.0.1")),
-        .package(url: "https://github.com/apple/swift-tools-support-core.git", branch: "main"),
     ]
 } else {
     package.dependencies += [
         .package(name: "SymbolKit", path: "../swift-docc-symbolkit"),
-        .package(path: "../swift-argument-parser"),
-        .package(path: "../swift-tools-support-core"),
     ]
 }
--- a/Package.swift
+++ b/Package.swift
@@ -26,6 +26,9 @@ let package = Package(
             capability: .command(
                 intent: .documentationGeneration()
             ),
+            dependencies: [
+                "snippet-build",
+            ],
             path: "Plugins/Swift-DocC Convert",
             exclude: ["Symbolic Links/README.md"]
         ),
@@ -74,7 +77,7 @@ if ProcessInfo.processInfo.environment["SWIFTCI_USE_LOCAL_DEPS"] == nil {
     package.dependencies += [
         .package(url: "https://github.com/apple/swift-docc-symbolkit", branch: "main"),
         .package(url: "https://github.com/apple/swift-argument-parser", .upToNextMinor(from: "1.0.1")),
-        .package(url: "https://github.com/apple/swift-tools-support-core.git", .branch("main")),
+        .package(url: "https://github.com/apple/swift-tools-support-core.git", branch: "main"),
     ]
 } else {
     package.dependencies += [
--- a/Package.swift
+++ b/Package.swift
@@ -51,6 +51,7 @@ let package = Package(
         .testTarget(
             name: "SwiftDocCPluginUtilitiesTests",
             dependencies: [
+                "Snippets",
                 "SwiftDocCPluginUtilities",
             ],
             resources: [
@@ -66,10 +67,12 @@ let package = Package(
             path: "Sources/SwiftDocCPluginDocumentation",
             exclude: ["README.md"]
         ),
+        .target(name: "Snippets"),
         .executableTarget(
             name: "snippet-build",
             dependencies: [
+                "Snippets",
                 .product(name: "SymbolKit", package: "swift-docc-symbolkit"),
-            ])
+            ]),
     ]
 )
--- a/Package.swift
+++ b/Package.swift
@@ -44,6 +44,9 @@ let package = Package(
                     description: "Preview the Swift-DocC documentation for a specified target."
                 )
             ),
+            dependencies: [
+                "snippet-build",
+            ],
             exclude: ["Symbolic Links/README.md"]
         ),
         
--- a/Package.swift
+++ b/Package.swift
@@ -20,6 +20,9 @@ let package = Package(
         .plugin(name: "Swift-DocC", targets: ["Swift-DocC"]),
         .plugin(name: "Swift-DocC Preview", targets: ["Swift-DocC Preview"]),
     ],
+    dependencies: [
+        .package(url: "https://github.com/apple/swift-docc-symbolkit", branch: "main"),
+    ],
     targets: [
         .plugin(
             name: "Swift-DocC",
@@ -70,13 +73,3 @@ let package = Package(
             ])
     ]
 )
-
-if ProcessInfo.processInfo.environment["SWIFTCI_USE_LOCAL_DEPS"] == nil {
-    package.dependencies += [
-        .package(url: "https://github.com/apple/swift-docc-symbolkit", branch: "main"),
-    ]
-} else {
-    package.dependencies += [
-        .package(name: "SymbolKit", path: "../swift-docc-symbolkit"),
-    ]
-}
--- a/Package.swift
+++ b/Package.swift
@@ -8,6 +8,7 @@
 // See https://swift.org/LICENSE.txt for license information
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 
+import Foundation
 import PackageDescription
 
 let package = Package(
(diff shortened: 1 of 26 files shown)


bitjammer:
Going ahead and opening this for referencing in an upcoming SE.

Package.swift
```diff
@@ -59,5 +63,20 @@ let package = Package(
...
         ),
+        .executableTarget(
+            name: "snippet-build",
+            dependencies: [
+                .product(name: "SymbolKit", package: "swift-docc-symbolkit"),
+            ])
     ]
 )
+
+if ProcessInfo.processInfo.environment["SWIFTCI_USE_LOCAL_DEPS"] == nil {
```
ethan-kusters, line 74: The Swift-DocC Plugin doesn't ship in the toolchain so we should be able to just add the dependencies directly for now.

Package.swift
```diff
@@ -59,5 +63,20 @@ let package = Package(
...
+    package.dependencies += [
+        .package(url: "https://github.com/apple/swift-docc-symbolkit", branch: "main"),
```
ethan-kusters, line 76: I think we'll need to figure out a release strategy for this. I suppose at the very least we can eventually update this to depend on the `release/5.7` branch once that is cut.

Sources/snippet-build/main.swift
```diff
@@ -0,0 +1,153 @@
...
+// Copyright (c) 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+
+import Foundation
+import SymbolKit
+
+struct SnippetBuildCommand {
```
ethan-kusters, line 12: Should we consider marking this `@main` and renaming the file to `SnippetBuildCommand`?

Sources/snippet-build/main.swift
```diff
@@ -0,0 +1,153 @@
...
+    var moduleName: String
+
+    static func printUsage() {
+        let usage = """
+            USAGE: snippet-build <snippet directory> <output directory> <module name>
+
+            ARGUMENTS:
+                <snippet directory> - The directory containing Swift snippets
+                <output directory> - The diretory in which to place Symbol Graph JSON file(s) representing the snippets
+                <module name> - The module name ot use for the Symbol Graph (typically should be the package name)
```
ethan-kusters, line 24:
```suggestion
                <module name> - The module name to use for the Symbol Graph (typically should be the package name)
```

Sources/snippet-build/main.swift
```diff
@@ -0,0 +1,153 @@
...
+            ARGUMENTS:
+                <snippet directory> - The directory containing Swift snippets
+                <output directory> - The diretory in which to place Symbol Graph JSON file(s) representing the snippets
+                <module name> - The module name to use for the Symbol Graph (typically should be the package name)
+            """
+        print(usage)
+    }
+
+    func run() throws {
+        print("Looking for snippets in \(snippetsDir.debugDescription)")
```
ethan-kusters, line 30: Should we gate this printing behind `--verbose` flag? Or reduce it a bit to better align with the existing SwiftPM output?

The current output seems a little busy:
```
Building for debugging...
Build complete! (0.14s)
Looking for snippets in "/Users/demo/Developer/swift-markdown-snippets/_Snippets"
Found 4 snippet groups, 7 snippets in total.
Writing snippet symbol graph to file:///Users/demo/Developer/swift-markdown-snippets/.build/plugins/Swift-DocC/outputs/.build/snippet-symbol-graphs/swift-markdown-swift-markdown-snippets/swift-markdown-snippets.symbols.json
Converting documentation...
Conversion complete! (1.06s)
Generated DocC archive at '/Users/demo/Developer/swift-markdown-snippets/.build/plugins/Swift-DocC/outputs/Markdown.doccarchive'
```

Sources/snippet-build/main.swift
bitjammer: Sure. It can make entry points a little harder to find but I don't expect there to be a ton of files here.

Package.swift
bitjammer: Yes. For whichever DocC version is in use, it should match its SymbolKit version.

Package.swift
bitjammer: Wouldn't this still be useful if the plugin builds with all of the other projects, even if it's not installed in the toolchain? Does the plugin only build in CI on its own?

ethan-kusters: Today the plugin only build in CI on itss own. I'd like to eventually get things set up so the plugin runs in SwiftPM's and Swift-DocC's CI but I'm not sure we need this even for that.

My understanding was `SWIFTCI_USE_LOCAL_DEPS` is only used for toolchain builds? But maybe I'm missing something there.

bitjammer: Well it sounds like it's not needed in any case, so I'll cut it.

Sources/snippet-build/main.swift
bitjammer: Since most folks will not be running this directly, I think probably we can just cut most of these print messages. I would like to keep the snippet count message, but since we don't have argument parser :/ probably let's just remove all of them for now.

Sources/snippet-build/SnippetBuildCommand.swift
```diff
@@ -0,0 +1,149 @@
...
+                <output directory> - The diretory in which to place Symbol Graph JSON file(s) representing the snippets
+                <module name> - The module name to use for the Symbol Graph (typically should be the package name)
+            """
+        print(usage)
+    }
+
+    func run() throws {
+        let snippetGroups = try loadSnippetsAndSnippetGroups(from: URL(fileURLWithPath: snippetsDir))
+
+        let totalSnippetCount = snippetGroups.reduce(0) { $0 + $1.snippets.count }
```
ethan-kusters, line 33: Just FYI- I'm getting a warning for this being unused.

```suggestion
```

Sources/snippet-build/SnippetBuildCommand.swift
```diff
@@ -0,0 +1,149 @@
...
+
+        guard !snippetGroups.isEmpty,
+              snippetGroups.allSatisfy({ !$0.snippets.isEmpty }) else {
```
ethan-kusters, line 36: Should we just filter to non-empty snippet groups here?

Sources/snippet-build/Snippets/Model/SnippetGroup.swift
```diff
@@ -0,0 +1,35 @@
...
+// Copyright (c) 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+
+import Foundation
+
+/// A group of snippets.
+public struct SnippetGroup {
```
ethan-kusters, line 12: I like the idea of putting the Snippet model into its own library actually but, since this is part of the executable, should it be public?

bitjammer: I'm basically preparing this to be its own library in general, and snippet groups will become more prominent in later use cases.

Sources/snippet-build/SnippetBuildCommand.swift
bitjammer: Just removed this actually.

Sources/snippet-build/SnippetBuildCommand.swift
bitjammer: This is to prevent an empty symbol graph file from showing up.

bitjammer:
@swift-ci Please test

Sources/snippet-build/SnippetBuildCommand.swift
ethan-kusters: Maybe I'm just misreading this. What about the case where one of the snippet groups is empty but not all of them?

Sources/Snippets/Parsing/PlainTextSnippetExtractor.swift
```diff
@@ -0,0 +1,140 @@
...
+        return trimmed
+    }
+
+    var isEmptyOrWhiteSpace: Bool {
+        return self.isEmpty || self.allSatisfy { $0.isWhitespace }
+    }
+}
+
+extension String {
+    mutating func removeLeadingAndTrailingNewlines() {
```
ethan-kusters, line 50: Is this equivalent to `.trimmingCharacters(in: .newlines)`?

Sources/snippet-build/SnippetBuildCommand.swift
bitjammer: Ah, you're right, there is a typo here. The logic should be:

- There need to be at least some snippet groups
- At least one snippet group needs to have a snippet

Which really means that the total snippet count needs to be > 0. I remember now why I added that :)

Sources/Snippets/Parsing/PlainTextSnippetExtractor.swift
```diff
@@ -0,0 +1,140 @@
...
+            }
+
+            if var comment = line.parsedLineCommentText,
+               comment.starts(with: "!") {
+                comment.removeFirst(1)
+                comment = comment.drop { $0.isWhitespace }
+                if lastExplanationLine.isEmptyOrWhiteSpace && comment.isEmptyOrWhiteSpace {
+                    continue
+                }
+                print(comment, to: &explanation)
```
ethan-kusters, line 121: The use of `print` here surprised me. Are there advantages/disadvantages here compared `.append(contentsOf:)`?

bitjammer: It's just a little nicer for automatic newlines, it's not strictly necessary.

Sources/Snippets/Parsing/PlainTextSnippetExtractor.swift
bitjammer: Yep!

ethan-kusters:
@swift-ci please test

ethan-kusters:
@swift-ci please test

ethan-kusters:
@swift-ci please test

ethan-kusters (approved):

Kyle-Ye:
@bitjammer Sorry to bother after this is merged.

Could we remove SymbolKit dependence on swift-docc-plugin since it is a basic plugin.

Once it depends on SymbolKit, it stops SymbolKit itself to use swift-docc-plugin. See https://github.com/apple/swift-docc-symbolkit/pull/19

bitjammer:
@Kyle-Ye I think it is not really possible to remove the SymbolKit dependency since that is how we will communicate snippet information to DocC, and we shouldn't manually generate JSON in this tool. I am investigating some other options in light of the dependency problem with SwiftPM plugins in general but the tool is already foregoing some useful dependencies.

PR #17, opened 2022-06-17
Snippets: Change show/hide markers to snippet.show and snippet.hide

bitjammer:
Per required changes to SE-0356, move the MARK: Show/Hide markers
to `snippet.show` and `snippet.hide`.

rdar://95220141

### DIFF
--- a/Sources/Snippets/Parsing/PlainTextSnippetExtractor.swift
+++ b/Sources/Snippets/Parsing/PlainTextSnippetExtractor.swift
@@ -13,7 +13,7 @@ enum SnippetVisibility {
 
 extension StringProtocol {
     /// If the string is a line comment, attempt to parse
-    /// a ``SnippetVisibility`` with `mark: show` or `mark: hide`.
+    /// a ``SnippetVisibility`` with `snippet.show` or `snippet.hide`.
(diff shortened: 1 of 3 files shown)


bitjammer:
@swift-ci Please test

Sources/Snippets/Parsing/PlainTextSnippetExtractor.swift
```diff
@@ -13,17 +13,17 @@ enum SnippetVisibility {
...
+    /// a ``SnippetVisibility`` with `snippet.show` or `snippet.hide`.
     var parsedVisibilityMark: SnippetVisibility? {
         guard var comment = parsedLineCommentText else {
             return nil
         }
 
         comment = comment.drop { $0.isWhitespace }
 
-        if comment.lowercased().starts(with: "mark: show") {
+        if comment.lowercased().starts(with: "snippet.show") {
```
denil-ct, line 24: Kinda curious, why starts with instead of checking for an exact match?

bitjammer: `snippet.show` is unlikely to be accidentally matched, but someone may add some comment text after the marker to justify hiding, or something else we haven't thought of. No need to be super strict here.

PR #21, opened 2022-07-20
Allow test script to use upstream DocC.

bitjammer:
Expect `SWIFT_DOCC_BRANCH` and `SWIFT_DOCC_RENDER_ARTIFACT_BRANCH` in the
environment. If both are set when running `./bin/test`, clone the respective
repos at those branches and set `DOCC_EXEC` and `DOCC_HTML_DIR` when running the
tests.

Exclude these cloned directories from `./bin/check-source`.

rdar://97294776

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -20,6 +20,41 @@ filepath() {
 # absolute file path to the Swift-DocC-Plugin root source dir.
 SWIFT_DOCC_PLUGIN_ROOT="$(dirname $(dirname $(filepath $0)))"
 
+SWIFT_DOCC_ROOT="$SWIFT_DOCC_PLUGIN_ROOT/swift-docc"
+SWIFT_DOCC_RENDER_ARTIFACT_ROOT="$SWIFT_DOCC_PLUGIN_ROOT/swift-docc-render-artifact"
+export DOCC_HTML_DIR="$SWIFT_DOCC_RENDER_ARTIFACT_ROOT/dist"
+
+SWIFT_DOCC_REPO=${SWIFT_DOCC_REPO:="https://github.com/apple/swift-docc.git"}
+SWIFT_DOCC_RENDER_ARTIFACT_REPO=${SWIFT_DOCC_RENDER_ARTIFACT_REPO:="https://github.com/apple/swift-docc-render-artifact.git"}
+
+SWIFT_DOCC_BRANCH=${SWIFT_DOCC_BRANCH:="main"}
+SWIFT_DOCC_RENDER_ARTIFACT_BRANCH=${SWIFT_DOCC_RENDER_ARTIFACT_BRANCH:="main"}
+
+# The script will clone swift-docc and swift-docc-render-artifact at the
+# branches pulled from the environment above.  The tests will then run using
+# that built DocC. This can be useful for testing interdependent changes that
+# need to land together and make it possible to test multiple pull requests
+# together.
+
+echo "Cloning docc..."
+rm -rf "$SWIFT_DOCC_ROOT"
+git clone -b "$SWIFT_DOCC_BRANCH" "${SWIFT_DOCC_REPO}" "$SWIFT_DOCC_ROOT" || exit 1
+
+echo "Cloning docc-render-artifact..."
+rm -rf "$SWIFT_DOCC_RENDER_ARTIFACT_ROOT"
+git clone -b "${SWIFT_DOCC_RENDER_ARTIFACT_BRANCH}" "${SWIFT_DOCC_RENDER_ARTIFACT_REPO}" "$SWIFT_DOCC_RENDER_ARTIFACT_ROOT" || exit 1
+
+echo "Building docc..."
+swift build --package-path "$SWIFT_DOCC_ROOT" --configuration release || exit 1
+
+export DOCC_EXEC="$(swift build --package-path "$SWIFT_DOCC_ROOT" --show-bin-path --configuration release)/docc"
+if [[ ! -f "$DOCC_EXEC" ]]; then
+  echo "docc executable not found, expected at $SWIFT_DOCC_EXEC"
+  exit 1
+else
+  echo "Using docc executable: $DOCC_EXEC"
+fi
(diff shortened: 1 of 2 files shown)


bin/test
```diff
@@ -20,6 +20,44 @@ filepath() {
...
+  git clone -b "$SWIFT_DOCC_BRANCH" "${SWIFT_DOCC_REPO}" "$SWIFT_DOCC_ROOT" || exit 1
+
+  echo "Cloning docc-render-artifact..."
+  rm -rf "$SWIFT_DOCC_RENDER_ARTIFACT_ROOT"
+  git clone -b "${SWIFT_DOCC_RENDER_ARTIFACT_BRANCH}" "${SWIFT_DOCC_RENDER_ARTIFACT_REPO}" "$SWIFT_DOCC_RENDER_ARTIFACT_ROOT" || exit 1
+
+  echo "Building docc..."
+  swift build --package-path "$SWIFT_DOCC_ROOT" --configuration release || exit 1
+
+  export DOCC_EXEC="$(swift build --package-path "$SWIFT_DOCC_ROOT" --show-bin-path --configuration release)/docc"
```
ethan-kusters, line 50: Ooh this is very slick!

bin/test
```diff
@@ -20,6 +20,44 @@ filepath() {
...
+SWIFT_DOCC_RENDER_ARTIFACT_REPO=${SWIFT_DOCC_RENDER_ARTIFACT_REPO:="ssh://git@github.com/apple/swift-docc-render-artifact.git"}
+
+# Both SWIFT_DOCC_BRANCH and SWIFT_DOCC_RENDER_ARTIFACT_BRANCH must be specified together
+# to prevent accidental incompatibilities.
+if [[ (-n "${SWIFT_DOCC_BRANCH:=}" && -z ${SWIFT_DOCC_RENDER_ARTIFACT_BRANCH:=}) || (-z "${SWIFT_DOCC_BRANCH:=}" && -n ${SWIFT_DOCC_RENDER_ARTIFACT_BRANCH:=}) ]]; then
+  echo "If either SWIFT_DOCC_BRANCH and SWIFT_DOCC_RENDER_ARTIFACT_BRANCH are specified, both must be specified."
+  exit 1
+fi
+
+if [[ ! -z "${SWIFT_DOCC_BRANCH:=}" ]]; then
```
ethan-kusters, line 36: Can we add a general comment somewhere around here explaining why were cloning and building docc?

bitjammer: No problem!

bin/test
bitjammer: It has come in handy many times now!

bitjammer:
@ethan-kusters Updated to clone `main` by default.

bitjammer:
@swift-ci Please test

bitjammer:
@swift-ci Please test

bitjammer:
Updated to use `https`.

bitjammer:
@swift-ci Please test

bin/test
```diff
@@ -20,6 +20,41 @@ filepath() {
...
+SWIFT_DOCC_RENDER_ARTIFACT_ROOT="$SWIFT_DOCC_PLUGIN_ROOT/swift-docc-render-artifact"
+export DOCC_HTML_DIR="$SWIFT_DOCC_RENDER_ARTIFACT_ROOT/dist"
+
+SWIFT_DOCC_REPO=${SWIFT_DOCC_REPO:="https://github.com/apple/swift-docc.git"}
+SWIFT_DOCC_RENDER_ARTIFACT_REPO=${SWIFT_DOCC_RENDER_ARTIFACT_REPO:="https://github.com/apple/swift-docc-render-artifact.git"}
+
+SWIFT_DOCC_BRANCH=${SWIFT_DOCC_BRANCH:="main"}
+SWIFT_DOCC_RENDER_ARTIFACT_BRANCH=${SWIFT_DOCC_RENDER_ARTIFACT_BRANCH:="main"}
+
+# The script will clone swift-docc and swift-docc-render-artifact at the
```
franklinsch, line 33: Can this be done optionally, when setting a command line flag or environment variable? There's still benefit in just using the `docc` that's in the user's toolchain; it makes running this script much faster.

bitjammer: When `./bin/test` is running, it is more likely to be someone either preparing for a PR or is actually running in PR CI. In both of those cases, it's not proper to run against the toolchain DocC–compatibility isn't guaranteed depending on which Xcode/toolchain is installed. This would be akin to downstream projects using the default toolchain's `swift` when they should be using the latest version.

If it becomes a situation where someone is running the tests over and over to hack on a problem and chip away at failing tests, contributors to this project can still run `swift test` and it will use the built-in DocC (again, not necessarily advisable), or a recently built DocC by using `DOCC_EXEC`.

If a command plugin relies on a tool built in another package, it might be useful in the future to be able to declare those kinds of dependencies in the manifest, which would cache build products, but that is still over the horizon in SwiftPM. If I heard correctly, that is something that they want to address in the future though.

ethan-kusters: I agree that running them by default makes sense (especially until we add support in CI for configuring this).

To Franklin's point though- maybe we should expose something like a `--skip-building-upstream-docc` flag for the case where the docc in the toolchain _is_ new enough? And then CI would still enforce that the plugin builds with the `main` DocC.

I don't feel too strongly either way though- if we want to land it like this and recommend using `swift test` directly for now, we can always open a separate PR in the future to add that flag if there's a need.

ethan-kusters (approved):
Thank you for adding this! I added an additional comment above by I'm happy to land this as is.

bin/test
franklinsch: `--skip-building-upstream-docc` would be nice, but not a blocker for this PR.

franklinsch (approved):

bin/test
bitjammer: Thanks. I think this would be a good first issue for someone in this project. https://github.com/apple/swift-docc-plugin/issues/22

PR #26, opened 2022-08-22
Preview: remove use of `signal` on Windows

compnerd:
Windows does not support signal(2).  Instead, use `TerminateProcess` to
terminate the process.  We will not properly handle the termination of
child processes for the preview process, which should be handled by the
construction of a job object that all child processes are attached to
such that when the parent process is terminated, the children will be
properly reaped.  This is required to enable building the plugin on
Windows which is a dependency for swift-syntax.

### DIFF
--- a/Plugins/Swift-DocC Preview/SwiftDocCPreview.swift
+++ b/Plugins/Swift-DocC Preview/SwiftDocCPreview.swift
@@ -6,6 +6,9 @@
 // See https://swift.org/LICENSE.txt for license information
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 
+#if os(Windows)
+import WinSDK
+#endif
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

compnerd:
CC: @DougGregor

QuietMisdreavus (commented):
This looks good to me, but i'll leave the final say to @ethan-kusters.

ethan-kusters (approved):
Thanks @compnerd! Is this blocking SwiftSyntax? I can tag a new release with just the fix cherry-picked if so.

CodaFi:
It's not blocking us, I'm going to temporarily remove this dependency from the main repository. Once we have a blue Windows build we can look into reintroducing it.

Issue #27, opened 2022-08-30
Support for skipping synthesized symbols


franklinsch:
#### Description:

The Swift compiler supports a `-skip-synthesized-members` argument which excludes synthesized symbols from symbol graph files. Synthesized symbols can lead to a large increase in number of documentation pages for some frameworks. We should have a way of generating documentation without synthesized symbols from the DocC plugin.

ethan-kusters:
This would be a great first issue! To support this we would need to add support for parsing a new flag and then change the boolean we're passing here: https://github.com/apple/swift-docc-plugin/blob/main/Plugins/SharedPackagePluginExtensions/Target%2BdefaultSymbolGraphOptions.swift#L29

Here's an example of a custom flag the plugin currently supports: https://github.com/apple/swift-docc-plugin/blob/main/Sources/SwiftDocCPluginUtilities/PluginFlags/DisableIndexFlag.swift#L14.

Happy to help with any questions if anyone would like to take this on!

natikgadzhi:
Was looking around for good small issues to get myself up to speed — looks like this one is complete in #28, @sofiaromorales 👏🏼 

@ethan-kusters, time to close this one? I would be happy to explore other issues and help out if I can, if you have anything that needs attention!

PR #28, opened 2022-09-19
Feat:  New plugin flag for skipping synthesized symbols

sofiaromorales:
Bug/issue #, if applicable: #27

## Summary

This pull request aims to support the generation of archives without synthesized symbols by passing a new plugin flag `--skip-synthesized-symbols`. This would reduce the size of the generated doccarchive.

The plugin flag `--skip-synthesized-symbols` has been added to achieve this. It doesn't transform any of the passed arguments but overrides the default value of `symbolGraphOptions.includeSynthesized` to false. Also, with the proposed implementation, more flags could be added to override other settings of the symbol graph options.

## Dependencies

N/A

## Testing

For automated testing run the integration tests suite.

For manual testing:
1. Download the attached package
2. In `Package.swift` set the right path for `swift-docc-plugin` dependency
3. Run `swift package generate-documentation`, and `swift package generate-documentation --skip-synthesized-symbols`, for both compare the files generated at `PackageWithConformanceSymbols.doccarchive/data/documentation/packagewithconformancesymbols/bar`

[PackageWithConformanceSymbols.zip](https://github.com/apple/swift-docc-plugin/files/9595998/PackageWithConformanceSymbols.zip)

## Checklist

- [X] Added tests
- [X] Ran the `./bin/test` script and it succeeded


IntegrationTests/Tests/Fixtures/PackageWithConformanceSymbols/Sources/PackageWithConformanceSymbols/FooTest.swift
```diff
@@ -0,0 +1,16 @@
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+
+import SwiftUI
```
ethan-kusters, line 9: I think it would be better to write a version of this test that doesn't depend on `SwiftUI` since `SwiftUI` is only available on Apple platforms.

Maybe conforming to a protocol in the Swift standard library, like [`Hashable`](https://developer.apple.com/documentation/swift/hashable) could work instead?

IntegrationTests/Tests/DocCConvertSynthesizedSymbolsTests.swift
```diff
@@ -0,0 +1,30 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+
+import XCTest
+
+final class DocCConvertSynthesizedSymbolsTests: XCTestCase {
+    func testGenerateDocumentationWithSkipSynthesizedSymbolsEnabled() throws {
+        let result = try swiftPackage(
```
ethan-kusters, line 13: I think we should have a second invocation of `swiftPackage` here without the `--skip-synthesized-symbols` flag and assert that the synthesized symbols do exist by default.

Sources/SwiftDocCPluginUtilities/PluginFlags/SkipSynthesizedSymbolsFlag.swift
```diff
@@ -0,0 +1,26 @@
...
+
+extension PluginFlag {
+    /// Excludes synthesized symbols from the generated documentation.
+    ///
+    /// If `--skip-synthesized-symbols` is found, pages for synthesized symbols
+    /// will not be generated, reducing the archive file size by not emitting any `view-implementations`
+    /// folder.
+    static let skipSynthesizedSymbols = PluginFlag(
+        parsedValues: [
+            "--skip-synthesized-symbols"
```
ethan-kusters, line 17: I think we should land this as an explicitly experimental feature to start to see how it impacts existing workflows before fully recommending it. For example, I expect existing DocC link completion strategies will now offer completions that aren't included in the built output when this flag is given.

```suggestion
            "--experimental-skip-synthesized-symbols"
```

Sources/SwiftDocCPluginUtilities/PluginFlags/SkipSynthesizedSymbolsFlag.swift
```diff
@@ -0,0 +1,26 @@
...
+        ],
+        abstract: "Excludes synthesized symbols from the generated documentation",
```
ethan-kusters, line 19:
```suggestion
        abstract: "Exclude synthesized symbols from the generated documentation",
```

Sources/SwiftDocCPluginUtilities/PluginFlags/SkipSynthesizedSymbolsFlag.swift
```diff
@@ -0,0 +1,26 @@
...
+        description: """
+            Produces a DocC archive with a reduced file size when converted by omitting synthesized symbols.
```
ethan-kusters, line 21: There are other reasons (besides reduced size on disk) that people might be interested in enabling this feature so I'm not sure it's worth mentioning that in the help text here.

```suggestion
            Experimental feature that produces a DocC archive without compiler synthesized symbols.
```

ethan-kusters (commented):
Looks good - thank you! I left some initial feedback and I'll want to do some more testing as well before we land this.

IntegrationTests/Tests/DocCConvertSynthesizedSymbolsTests.swift
sofiaromorales: Done 👍 Added tests for both cases, with and without `--experimental-skip-synthesized-symbols`.

IntegrationTests/Tests/Fixtures/PackageWithConformanceSymbols/Sources/PackageWithConformanceSymbols/FooTest.swift
sofiaromorales: Thanks for pointing this out. I removed the `SwiftUI` conformance and changed it to `Hashable` instead. Resolved here https://github.com/apple/swift-docc-plugin/pull/28/commits/b20147a2dab588c70a7b1d54a028d3132e01a86f.

Sources/SwiftDocCPluginUtilities/PluginFlags/SkipSynthesizedSymbolsFlag.swift
sofiaromorales: Makes sense, I changed the naming to ` "--experimental-skip-synthesized-symbols"`

Sources/SwiftDocCPluginUtilities/PluginFlags/SkipSynthesizedSymbolsFlag.swift
sofiaromorales: 👍 https://github.com/apple/swift-docc-plugin/pull/28/commits/b20147a2dab588c70a7b1d54a028d3132e01a86f

Sources/SwiftDocCPluginUtilities/PluginFlags/SkipSynthesizedSymbolsFlag.swift
sofiaromorales: Thanks, Ethan! 🙏 https://github.com/apple/swift-docc-plugin/pull/28/commits/b20147a2dab588c70a7b1d54a028d3132e01a86f

ethan-kusters:
Hi @sofiaromorales would you mind taking a look at the merge conflicts here when you have a chance? I believe #34 introduced a few since it also touches some of this same code.

It would be great to land this in the next release alongside extension support if we can.

Sources/SwiftDocCPluginUtilities/ParsedArguments.swift
```diff
@@ -201,7 +201,8 @@ private extension ParsedArguments {
             switch self {
             case .dumpSymbolGraph:
                 return [
-                    PluginFlag.extendedTypes
+                    PluginFlag.extendedTypes,
+                    PluginFlag.skipSynthesizedSymbols
```
sofiaromorales, line 205: I followed along the same strategy as in #34 since`--experimental-skip-synthesized-symbol` is transformed to `-skip-synthesized-members` flag of the `swift package dump-symbol-graph` command

ethan-kusters:
@swift-ci please test

Tests/SwiftDocCPluginUtilitiesTests/HelpInformationTests.swift
```diff
@@ -21,6 +21,9 @@ final class HelpInformationTests: XCTestCase {
             doccExecutableURL: URL(fileURLWithPath: "/")
         )
         
+        print("includeExtendedTypesSection")
```
ethan-kusters, line 24: Should these `print`s be removed?

Tests/SwiftDocCPluginUtilitiesTests/ParsedArgumentsTests.swift
```diff
@@ -455,15 +455,14 @@ final class ParsedArgumentsTests: XCTestCase {
             outputPath: "/my/output-path"
         )
         
-        XCTAssertFalse(doccArguments.contains("--include-extended-types"))
```
ethan-kusters, line 458: Should we add back this assertion?

Sources/SwiftDocCPluginUtilities/ParsedArguments.swift
```diff
@@ -180,9 +174,15 @@ public struct ParsedArguments {
         "--index",
     ]
     
+    // Build array with plugin flags that modify the symbol graph generation,
+    // filtering from the available custom symbol graph options those
+    // that correspond to the received flags
+    var symbolGraphArguments: [PluginFlag]
```
ethan-kusters, line 180: Is there a need to move this declaration? If not – I think it would be better to keep the diff more focused by leaving it on line 164.

Tests/SwiftDocCPluginUtilitiesTests/PluginFlags/PluginFlagTests.swift
```diff
@@ -76,4 +76,34 @@ final class PluginFlagTests: XCTestCase {
...
+            argumentTransformation: { $0 }
+        )
+        let examplePluginFlag4 = PluginFlag(
+            parsedValues: ["--example"],
+            abstract: "",
+            description: "",
+            argumentTransformation: { $0 }
+        )
+        XCTAssertFalse(examplePluginFlag1 == examplePluginFlag2)
+        XCTAssertTrue(examplePluginFlag1 == examplePluginFlag3)
```
ethan-kusters, line 106: Nit: We'll see more helpful error messages if we use:

```suggestion
        XCTAssertNotEqual(examplePluginFlag1, examplePluginFlag2)
        XCTAssertEqual(examplePluginFlag1, examplePluginFlag3)
```

IntegrationTests/Tests/DocCConvertSynthesizedSymbolsTests.swift
```diff
@@ -0,0 +1,49 @@
...
+//
+// Copyright (c) 2022 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+
+import XCTest
+
+final class DocCConvertSynthesizedSymbolsTests: XCTestCase {
```
ethan-kusters, line 11:
```suggestion
final class DocCConvertSynthesizedSymbolsTests: ConcurrencyRequiringTestCase {
```

This should fix the test failures on macOS. We currently don't have a Toolchain that supports Swift concurrency on the macOS CI.

ethan-kusters:
CC: @theMomax I'm having trouble adding you as reviewer for some reason – but would appreciate your review on this as well.

theMomax:
@ethan-kusters that's probably because I don't have commit access to Swift repositories and thus cannot formally approve a PR.

As for the review, I think it looks good! One thing I didn't check, though, is whether the `symbolGraphOptions.includeSynthesized` property has existed since 5.6. I do think so, but otherwise we'd need some special handling for backwards compatibility as I added it in #34.

ethan-kusters:
@swift-ci please test

ethan-kusters (approved):
Looks great. Thank you @sofiaromorales!

PR #33, opened 2022-10-22
Rename snippet-build to snippet-extract

bitjammer:
In general, the "build" terminology isn't really correct for snippets, as no compilation is going on as far as the plugin is concerned. This tool is actually parsing and extracting text for creation of symbol graphs for a package's snippets. So we should call this process "extraction" rather than "building".

rdar://101436368

## Checklist

- [x] NFC, no change to tests except for renaming.


bitjammer:
@swift-ci Please test

bitjammer:
@swift-ci Please test

ethan-kusters (approved):

PR #34, opened 2022-11-13
Add --include-extended-types flag

theMomax:
Bug/issue #, if applicable: apple/swift-docc#210

## Summary

The `--include-extended-types` flag is transformed to the `--emit-extension-block-symbols` flag of the `swift package dump-symbol-graph` command. It is available starting from Swift 5.8.

## Dependencies

All dependencies have been merged. For an overview see https://github.com/apple/swift-docc/issues/210#issuecomment-1312692921.

## Testing

1. Use a toolchain generated from `main` or `release/5.8` branches after December 21st
2. Use the base branch of this PR when adding the `swift-docc-plugin` to your `Package.swift`:
```swift
dependencies: [
    .package(url: "https://github.com/themomax/swift-docc-plugin", branch: "add-extended-types-flag")
]
```
3. Add the `--include-extended-types` flag when running your command

Verifying the basic functionality can be done with [this](https://github.com/theMomax/swift-docc-sample-nested-types) sample repository. Running `swift package --disable-sandbox preview-documentation --include-extended-types --target NestedTypes` should produce the result shown below:

image

## Checklist

- [x] Added tests
- [x] Ran the `./bin/test` script and it succeeded:
Swift 5.7.2: `.bin/test` succeeded
Swift 5.8/main nightly: `.bin/test` succeeded after removing the `--parallel` flag from `swift test` invocations in the script (Running `swift test --parallel` on any package fails for me with an error similar to the one described [here](https://github.com/apple/swift-package-manager/issues/4562). However, this is an issue with the nightly toolchains I've been getting for a while now.)


theMomax:
@ethan-kusters this PR is ready for review now!

Plugins/SharedPackagePluginExtensions/SymbolGraphOptions+override.swift
```diff
@@ -0,0 +1,28 @@
...
+        for argument in arguments {
+            switch argument {
+            case "--emit-extension-block-symbols":
+#if swift(>=5.8)
+                self.emitExtensionBlocks = true
+#else
+                print("warning: detected '--emit-extension-block-symbols' option, which is incompatible with your swift version (required: 5.8)")
+#endif
+            default:
+                print("warning: detected unknown dump-symbol-graph option '\(argument)'")
```
ethan-kusters, line 24: Should we `fatalError` here? This is a programmer error, right? It shouldn't be possible to receive an unknown flag here?

Plugins/SharedPackagePluginExtensions/Target+defaultSymbolGraphOptions.swift
```diff
@@ -27,7 +27,21 @@ extension SwiftSourceModuleTarget {
...
+            includeSPI: false,
+            emitExtensionBlocks: false
         )
     }
 }
+
+
+#if swift(<5.8)
+private extension PackageManager.SymbolGraphOptions {
+    /// A copatibility layer for lower Swift versions which discards unknown parameters.
```
ethan-kusters, line 39:
```suggestion
    /// A compatibility layer for lower Swift versions which discards unknown parameters.
```

Plugins/SharedPackagePluginExtensions/Target+defaultSymbolGraphOptions.swift
```diff
@@ -27,7 +27,21 @@ extension SwiftSourceModuleTarget {
...
-            includeSPI: false
+            includeSPI: false,
+            emitExtensionBlocks: false
         )
     }
 }
+
+
+#if swift(<5.8)
+private extension PackageManager.SymbolGraphOptions {
```
ethan-kusters, line 38: Have you tested the version of the plugin built in this PR the current Mac App Store Xcode?

Sources/SwiftDocCPluginUtilities/ParsedArguments.swift
```diff
@@ -170,6 +185,51 @@ public struct ParsedArguments {
...
+            case .docc:
+                return nil
+            }
+        }
+    }
+}
+
+private extension Arguments {
+    /// Returns the subset of arguments which are applicable to the given `consumer`.
+    func filter(for consumer: ParsedArguments.ArgumentConsumer) -> Arguments {
```
ethan-kusters, line 219: Nice 👍🏻

Sources/SwiftDocCPluginUtilities/ParsedArguments.swift
```diff
@@ -19,10 +19,25 @@ public struct ParsedArguments {
         return arguments.contains("--help") || arguments.contains("-h")
     }
     
+    /// Returns the arguments that could be passed to `swift package dump-symbol-graph`.
+    ///
+    /// In practice, however we won't invoke the `dump-symbol-graph` command via the command line,
```
ethan-kusters, line 24: Should we consider a different way to represent this information instead of strings representing the flags since we're going to re-parse them anyway?

I think tying this implementation to `swift package dump-symbol-graph` is kind of unnecessarily confusing for folks who might be familiar with SwiftPM plugins but not that specific command.

Sources/SwiftDocCPluginUtilities/ParsedArguments.swift
```diff
@@ -170,6 +185,51 @@ public struct ParsedArguments {
...
+    /// Returns the subset of arguments which are applicable to the given `consumer`.
+    func filter(for consumer: ParsedArguments.ArgumentConsumer) -> Arguments {
+        if let flagsToInclude = consumer.flags {
+            return self.filter { argument in
+                flagsToInclude.contains(where: { flag in
+                    flag.parsedValues.contains(argument)
+                })
+            }
+        } else {
+            let flagsToExclude = ParsedArguments.ArgumentConsumer.allCases.compactMap(\.flags).flatMap { $0 }
```
ethan-kusters, line 227: Could you add a couple of comments to this method body explaining the implementation/rationale? It took me a little while to fully understand what was going on here.

Sources/SwiftDocCPluginUtilities/PluginFlags/ExtendedTypesFlag.swift
```diff
@@ -0,0 +1,29 @@
...
+    /// Enables the extension block symbol format when calling the
+    /// dump symbol graph API.
+    ///
+    /// - Note: This flag is only available starting from Swift 5.8. It should
+    /// be hidden from the `--help` command for lower toolchain versions.
+    /// However, we do not hide the flag entirely, because this enables us to give
+    /// a more precise warning when accidentally used with Swift 5.7 or lower.
+    static let extendedTypes = PluginFlag(
+        parsedValues: [
+            "--include-extended-types",
```
ethan-kusters, line 21: Do you think we should consider adding an `--exclude-extended-types` flag as part of this PR? I think we'll need one in the future regardless but I could either way on adding it now or later.

ethan-kusters (commented):
Looks great! Just added a few initial thoughts.

ethan-kusters:
@swift-ci please test

Plugins/SharedPackagePluginExtensions/SymbolGraphOptions+override.swift
theMomax: For the incompatible version I'd prefer to do it the same way as in SwiftPM, i.e. emitting a warning (see https://github.com/apple/swift-package-manager/pull/5892#issuecomment-1352727491).

The typed configuration options you suggested make the default case for unknown flags redundant anyway.

Plugins/SharedPackagePluginExtensions/Target+defaultSymbolGraphOptions.swift
theMomax: Yes, of course! I tested with the latest release as well as with recent nightly versions.

For the release version I tried Xcode `14.2` with the default toolchain for building and `swift-driver version: 1.62.15 Apple Swift version 5.7.2 (swiftlang-5.7.2.135.5 clang-1400.0.29.51)` when running the plugin in the command-line.

Sources/SwiftDocCPluginUtilities/ParsedArguments.swift
theMomax: I think that might be a good idea. I gave it a try based on the following protocol:

```swift
/// Transforms a set of arguments, transferring information to a typed configuration.
protocol ArgumentsTransformer: ArgumentsTransforming {
    associatedtype Configuration
    /// Apply the transformation to the given arguments and configuration.
    func transform(arguments: inout Arguments, configuration: inout Configuration)
}
```

Please take a look and let me know what you think about it. I'm not super happy with the result, but I'm also not really happy with the whole CLI parsing setup in this plugin. I'm also wondering if a ArgumentParser based rewrite wouldn't be favorable in general.

Sources/SwiftDocCPluginUtilities/ParsedArguments.swift
theMomax: Sure :)

Sources/SwiftDocCPluginUtilities/PluginFlags/ExtendedTypesFlag.swift
theMomax: This is actually one of the points where things get really tricky (or ugly) with the current CLI parsing setup. With the current setup, each `PluginFlag` iterates over the whole (remaining) list of `Arguments` individually. With the order in which the `PluginFlag`s are applied is defined somewhat arbitrary and constant. Now, if I were to implement a second `PluginFlag` for the `--exclude-extended-types` flag, that would cause a problem when the user tries to override flags:

I assume (without loss of generality), that the _include_ `PluginFlag` is applied first and the _exclude_ version second. Now take the following example:

```bash
swift package generate-documentation --exclude-extended-types --include-extended-types
```

Other than what one would expect, extension block symbols will be disabled, because - independently of the order of CLI flags - the _exclude_ `PluginFlag` is executed last and will always see the `--exclude-extended-types` flag.

I could fix this problem by making one `PluginFlag` for both `--exclude-extended-types` and `--include-extended-types`, but then I'd somehow have to pass the matched flags into the `transformation`, which would make the whole `PluginFlag` abstraction somewhat redundant.

-----------------------

I'd therefore suggest to keep that part out of this PR so we can have a lengthier discussion about the long-term goal for the plugin's CLI parsing system before we invest more time into it.

Sources/SwiftDocCPluginUtilities/ParsedArguments.swift
ethan-kusters: Looks great – thank you!

Sources/SwiftDocCPluginUtilities/PluginFlags/ExtendedTypesFlag.swift
ethan-kusters: That sounds reasonable to me. 👍🏻

Sources/SwiftDocCPluginUtilities/HelpInformation.swift
```diff
@@ -43,17 +43,19 @@ public enum HelpInformation {
...
             PluginFlag.disableIndex,
         ]
         
+        // stops 'not mutated' warning for Swift 5.7 and lower
+        supportedPluginFlags += []
+        
+#if swift(>=5.8)
+        supportedPluginFlags += [PluginFlag.extendedTypes]
+#endif
+
```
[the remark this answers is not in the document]
theMomax: No, unfortunately we're limited to statements here :(

Plugins/SharedPackagePluginExtensions/SymbolGraphOptions+override.swift
theMomax: I added the fatalError [here](https://github.com/apple/swift-docc-plugin/blob/e71e7c38a126e59be3d496d7aed3bd760b1f9633/Plugins/SharedPackagePluginExtensions/PackageManager%2BgetSymbolGraphsForDocC.swift#L59), as it becomes relevant again with the approach from #28

Sources/SwiftDocCPluginUtilities/ParsedArguments.swift
theMomax: > This is 100% what I'd like to do long-term. We're just blocked on SwiftPM support to allow command plugins to have dependencies: https://github.com/apple/swift-package-manager/issues/5574.

Ah, yeah, I forgot about that :)

----------

> I'm worried this approach with the generic Configuration is over-optimizing for what we need here. I'm curious if you've looked at the approach in https://github.com/apple/swift-docc-plugin/pull/28? It's less extensible but since we're likely going to need to overhaul our argument parsing solution here in the long-term anyway I wonder if something along those lines would be more maintainable in the shorter term.

I felt so too, and, now that I've looked at #28, I agree it's the better short-term solution. I adopted the approach, making sure to cause as little merge-conflicts as possible. I only kept the `filter(for:)` function in `ParsedArguments` from my approach, as I felt like without it it would be too easy to accidentally have flags passed to `docc` that are not meant for it.

Sources/SwiftDocCPluginUtilities/HelpInformation.swift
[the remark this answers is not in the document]
ethan-kusters: Sad 😢

ethan-kusters (commented):
Looks great! Thank you so much for your work here – really excited to get this in.

Could we add a couple integration tests as well before landing this? I put together a patch that I think gives enough basic coverage but feel free to expand from there: 

[add-integration-test.patch](https://github.com/apple/swift-docc-plugin/files/10408687/add-integration-test.patch)

```
git apply add-integration-test.patch
```

You can open up the sub-package in the `swift-docc-plugin/IntegrationTests` subdirectory to run the tests. (I think the parent package needs to be closed for things to work.)

theMomax:
@ethan-kusters I expanded your test with a protocol conformance, but I don't think it makes sense to go any further. After all, the tests here should just verify that the plugin invokes the toolchain correctly, which can easily be verified with these simple cases. Anything beyond that should be/is tested in `swift-docc` and `libSymbolGraphGen` in the compiler.

ethan-kusters:
@swift-ci please test

ethan-kusters:
@swift-ci please test

ethan-kusters (approved):
Looks great! Thank you! 🚀

ethan-kusters:
@swift-ci please test

daniel-grumberg:
@swift-ci please test

PR #40, opened 2022-12-09
Disable integration tests on macOS to unblock CI

ethan-kusters:
SwiftPM command plugins can only be run when the Swift concurrency libraries are available. For some reason these libraries are no longer available in the macOS used in SwiftCI so this disables those tests after performing a check to see if concurrency is available.

This mirrors the implementation in SwiftPM itself here: https://github.com/apple/swift-package-manager/blob/dd7e9cc6/Sources/SPMTestSupport/Toolchain.swift#L55

### DIFF
--- a/IntegrationTests/Tests/DocCArchiveIndexGenerationTests.swift
+++ b/IntegrationTests/Tests/DocCArchiveIndexGenerationTests.swift
@@ -8,7 +8,7 @@
 
 import XCTest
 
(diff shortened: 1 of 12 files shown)


ethan-kusters:
@swift-ci please test

ethan-kusters:
@swift-ci please test

ethan-kusters:
@swift-ci please test

bitjammer (approved):
Looks good!

PR #42, opened 2023-01-09
Only build the required `docc` executable product in CI

ethan-kusters:
Resolves an issue where CI is failing with:

```
/swift-docc-plugin/swift-docc/Sources/generate-symbol-graph/main.swift:13:18: error: module 'SwiftDocC' was not compiled for testing
@testable import SwiftDocC
~~~~~~~~~~       ^
```

Caused by: https://github.com/apple/swift-docc/pull/447

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -49,11 +49,11 @@ then
   git clone -b "${SWIFT_DOCC_RENDER_ARTIFACT_BRANCH}" "${SWIFT_DOCC_RENDER_ARTIFACT_REPO}" "$SWIFT_DOCC_RENDER_ARTIFACT_ROOT" || exit 1
 
   echo "Building docc..."
-  swift build --package-path "$SWIFT_DOCC_ROOT" --configuration release || exit 1
+  swift build --package-path "$SWIFT_DOCC_ROOT" --product docc --configuration release || exit 1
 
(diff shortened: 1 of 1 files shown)


ethan-kusters:
@swift-ci please test

bitjammer:
Change looks good. For posterity, can you explain how limiting the CI build to just the `docc` product fixes this issue?

bitjammer (approved):

ethan-kusters:
> Change looks good. For posterity, can you explain how limiting the CI build to just the docc product fixes this issue?

I think we're unable to build the `generate-symbol-graph` product in release mode by default because it relies on a `@testable import`. Potentially (https://github.com/apple/swift-package-manager/issues/5071) we could add another flag to resolve this but since the `@testable` import was really a temporary workaround while we stabilize the API `generate-symbol-graph` is depending on I think this is a reasonable solution here.

bitjammer:
Thanks for explaining!

bitjammer:
Curious how this also works if specifying that `SWIFT_SKIP_BUILDING_UPSTREAM_DOCC` flag we added some time back?

ethan-kusters:
> Curious how this also works if specifying that `SWIFT_SKIP_BUILDING_UPSTREAM_DOCC` flag we added some time back? I know that doesn't affect CI, just wondering.

Tests should pass when providing that flag, right? I haven't actually tried it...

bitjammer:
Yep, thanks.

PR #43, opened 2023-01-10
Copy snippet symbol graphs to the correct final destination

bitjammer:
The snippet symbol graph is currently being generated with the changes in e013865138b, however they are not being copied to the correct final destination.

The `snippet-extract` tool now takes input and output *files* instead of directories, however there is still one place in the plugin where we are still treating the output like a directory, so the snippet symbol graph basename path component looks like a directory, and needs additional basenames to put it in the right place with the correct file extension.

rdar://104045850

- [x] Added tests


bitjammer:
Thank you @ethan-kusters for the patch!

bitjammer:
@swift-ci Please test

bitjammer:
@swift-ci Please test

ethan-kusters:
@swift-ci please test

PR #49, opened 2023-01-27
Create .spi.yml

finestructure:
As discussed with @franklinsch , this will ensure SPI documentation is updated without any particular workarounds in place in SPI.

We initially added custom support to host documentation on SPI without a `.spi.yml` file but this is becoming a bit tricky to maintain and has broken in a couple of instances.

### DIFF
--- a/.spi.yml
+++ b/.spi.yml
@@ -1,0 +1,4 @@
+version: 1
+builder:
+  configs:
+    - documentation_targets: [SwiftDocCPlugin]


.spi.yml
```diff
@@ -0,0 +1,4 @@
+version: 1
+builder:
+  configs:
+    - documentation_targets: [SwiftDocCPlugin]
```
finestructure, line 4: There are a couple of other targets in the package but I believe this is the one we'd configured in our override. Or should there be others in this list as well?

franklinsch: @ethan-kusters can you confirm?

ethan-kusters: That's correct 👍🏻

ethan-kusters (approved):

ethan-kusters:
@swift-ci please test

PR #61, opened 2023-06-22
Replace os() with canImport(Darwin)

brennobemoura:
## Summary

_Provide a description of what your PR addresses, explaining the expected user experience. 

I updated the `#if os()` usage for Apple's platform to a better solution as `#if canImport(Darwin)`. With this solution, there's no need to add one more `os(xrOS)`.

## Checklist

I didn't implement anything new, so no documentation is needed to be updated neither implementing additional tests.


brennobemoura:
As discussed [here](https://github.com/apple/swift-collections/pull/298) this solution may not fit well. Since Swift doesn’t isolate packages based on their dependency tree, this code can actually introduce new bugs.

So I'll close this for now, as I understand more about the problem I was trying to solve.

PR #69, opened 2023-12-11
bump to current version in example snippet

dastrobu:
## Summary

bump to current version in example snippet## Dependencies

## Checklist

- [ ] Added tests 
   docs only 
- [ ] Ran the `./bin/test` script and it succeeded
   docs only 

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ let package = Package(
     // name, platforms, products, etc.
     dependencies: [
         // other dependencies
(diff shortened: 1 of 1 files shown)


heckj:
@dastrobu this isn't strictly needed, since the `from:` stanza in the dependency tells SwiftPM (or Xcode) to get resolve with this as a minimum version, up to the next major semantic version. So adding in updates to the README for each of the version increments wouldn't make using it as an example value in a real project any more or less effective.

PR #103, opened 2024-12-22
add `.vscode` to .gitignore

RandomHashTags:
Bug/issue #, if applicable: N/A

## Summary

This PR adds the `.vscode` to the `.gitignore` file to exclude vs code configuration files from contributors using VS Code.

## Dependencies

N/A

## Testing

N/A

## Checklist

- [x] <s>Added tests</s> Unit testing unnecessary
- [x] <s>Updated documentation if necessary</s> Documenting this change is unnecessary

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ DerivedData/
 .swiftpm
 /gh-pages
 Package.resolved
+.vscode


heckj:
Thanks @RandomHashTags 

While we don't have a `.vscode` directory here today, that's something 
we don't want to exclude within `.gitignore`. Rather than adding it as an explicit ignore to the project, we'd encourage you to add it `.git/info/exclude`. That's a local-to-your-repository configuration that'll achieve the same goal, without impacting the project more generally.

PR #120, opened 2026-04-03
Support snippets in files other than .swift

### LINKED ISSUE #1436: The ability to include source content (markdown, or programming sources) from outside the catalog (beyond Swift snippets)
### Feature Name

snippets-beyond-swift

### Description

I'd like to have an easy path so that content that is stored within a Swift package can be easily referenced and used within the content presented in a DocC catalog. 

### Motivation

The two primarily use cases here are markdown files that aren't housed within the catalog due to overlap and value in their location within a broader project structure, and including source code from languages other than Swift, which is all the snippets provides today.

For markdown files, being able to reference and/or include the README while leaving it at the root of the git repo (typically VERY outside of a catalog) has been an extremely common request. And this just as easily applies to generated sources that might be part of a build process that aren't easy to incorporate into DocC content documentation today.

The source file desire comes from wanting to lean into the broader interop stories for Swift - specifically C and C++, but it applies to any other possible language, where today it's impossible to reference a validated, tested source file - or a slice of such a file, akin to how Snippets allows you to slice a Swift file with comments - in a DocC article. Being able to define something within a catalog that points to a specific source file, and use that as a file that can be converted into a "snippet" for inclusion within any of the DocC catalogs that may be supported within a project, would open up documenting how to support interop with other languages significantly from where we are today - which relies on the high maintenance "copy the snippet into place".

Additionally, something beyond the snippets directory would support allowing a more direct "here's how this code works" linkage, referencing the source files and displaying portions of them inline with content, allowing more contributor-focused documentation to be able to be created.

### Importance

Reduces the friction for managing source code snippets, particularly for non-Swift languages, and include their content in DocC articles.

### Alternatives Considered

Manually copying and maintaining a replica of any relevant source code into a DocC catalog.

ktoso:
This resolves https://github.com/swiftlang/swift-docc/issues/1436

## Summary

This removes the restriction of only scanning .swift files and we try to detect the language based on the extension. If a language is not known, we pass the extension instead as a fallback -- this should be acceptable for most cases.

This is important to unblock swift-java documenting its feature set with compiling examples, like this:
Screenshot 2026-04-03 at 12 56 34

## Testing

Added tests

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/Snippets/Model/SnippetLanguage.swift
+++ b/Sources/Snippets/Model/SnippetLanguage.swift
@@ -6,11 +6,30 @@
 // See https://swift.org/LICENSE.txt for license information
 // See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 
-/// A programming language supported for code snippets.
-///
-/// Only languages that use `//` line comments are currently supported,
-/// since ``SnippetParser`` uses `//` as the comment prefix for snippet
-/// markers and explanations.
+/// The comment syntax used by a language for snippet markers and explanations
+public enum CommentStyle: Sendable {
+    /// Line comment with a given prefix, e.g. `//` or `#`
+    case lineComment(String)
+    /// Block comment with a prefix and suffix, e.g. `<!--` and `-->`
+    case blockComment(prefix: String, suffix: String)
+
+    /// Inline comment style using `//`
+    public static var slashSlash: Self { 
+        .lineComment("//")
+    }
+    
+    /// Inline comment style using `#`
+    public static var hash: Self { 
+        .lineComment("#")
+    }
+
+    /// Default comment style for XML/HTML comments
+    public static var xml: Self {
+        .blockComment(prefix: "<!--", suffix: "-->")
+    }
+}
+
+/// A programming language supported for code snippets
 public enum SnippetLanguage: String, CaseIterable, Sendable {
     case swift
     case java
@@ -25,6 +44,11 @@ public enum SnippetLanguage: String, CaseIterable, Sendable {
     case typescript
     case scala
     case groovy
+    case python
+    case bash
+    case zsh
+    case xml
+    case html
 
     /// The language identifier used in symbol graphs and for syntax highlighting
     public var id: String {
@@ -39,9 +63,26 @@ public enum SnippetLanguage: String, CaseIterable, Sendable {
         case .go:         return "go"
         case .rust:       return "rust"
         case .javascript: return "javascript"
-        case .typescript:  return "typescript"
+        case .typescript: return "typescript"
         case .scala:      return "scala"
         case .groovy:     return "groovy"
+        case .python:     return "python"
+        case .bash:       return "bash"
+        case .zsh:        return "zsh"
+        case .html:       return "html"
+        case .xml:        return "xml"
+        }
+    }
+
+    /// The comment style used by this language
+    public var commentStyle: CommentStyle {
+        switch self {
+        case .python, .bash, .zsh:
+            return .hash
+        case .xml, .html:
+            return .xml
+        default:
+            return .slashSlash
         }
     }
 
@@ -61,6 +102,11 @@ public enum SnippetLanguage: String, CaseIterable, Sendable {
         case .typescript:  return ["ts", "tsx"]
         case .scala:       return ["scala"]
         case .groovy:      return ["groovy"]
+        case .python:      return ["py"]
+        case .bash:        return ["sh", "bash"]
+        case .zsh:         return ["zsh"]
+        case .html:        return ["html", "htm", "xhtml"]
+        case .xml:         return ["xml"]
         }
     }
 
@@ -76,5 +122,13 @@ public enum SnippetLanguage: String, CaseIterable, Sendable {
     public static func language(forFileExtension ext: String) -> SnippetLanguage? {
         let lowered = ext.lowercased()
         return allCases.first { $0.fileExtensions.contains(lowered) }
+    }
+
+    /// Look up the comment style for a given file extension
+    ///
+    /// Returns `.slashSlash` for unknown extensions, since most programming
+    /// languages use `//` line comments
+    public static func commentStyle(forFileExtension ext: String) -> CommentStyle {
+        language(forFileExtension: ext)?.commentStyle ?? .slashSlash
     }
 }
--- a/Sources/Snippets/Model/SnippetLanguage.swift
+++ b/Sources/Snippets/Model/SnippetLanguage.swift
@@ -1,0 +1,80 @@
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+
+/// A programming language supported for code snippets.
+///
+/// Only languages that use `//` line comments are currently supported,
+/// since ``SnippetParser`` uses `//` as the comment prefix for snippet
+/// markers and explanations.
+public enum SnippetLanguage: String, CaseIterable, Sendable {
+    case swift
+    case java
+    case kotlin
+    case c
+    case cpp
+    case objectiveC
+    case csharp
+    case go
+    case rust
+    case javascript
+    case typescript
+    case scala
+    case groovy
+
+    /// The language identifier used in symbol graphs and for syntax highlighting
+    public var id: String {
+        switch self {
+        case .swift:      return "swift"
+        case .java:       return "java"
+        case .kotlin:     return "kotlin"
+        case .c:          return "c"
+        case .cpp:        return "c++"
+        case .objectiveC: return "objective-c"
+        case .csharp:     return "csharp"
+        case .go:         return "go"
+        case .rust:       return "rust"
+        case .javascript: return "javascript"
+        case .typescript:  return "typescript"
+        case .scala:      return "scala"
+        case .groovy:     return "groovy"
+        }
+    }
+
+    /// The file extensions associated with this language
+    public var fileExtensions: [String] {
+        switch self {
+        case .swift:       return ["swift"]
+        case .java:        return ["java"]
+        case .kotlin:      return ["kt"]
+        case .c:           return ["c", "h"]
+        case .cpp:         return ["cpp", "hpp", "cc", "cxx"]
+        case .objectiveC:  return ["m", "mm"]
+        case .csharp:      return ["cs"]
+        case .go:          return ["go"]
+        case .rust:        return ["rs"]
+        case .javascript:  return ["js", "jsx"]
+        case .typescript:  return ["ts", "tsx"]
+        case .scala:       return ["scala"]
+        case .groovy:      return ["groovy"]
+        }
+    }
+
+    /// All file extensions supported for snippet extraction
+    public static var supportedFileExtensions: Set<String> {
+        Set(allCases.flatMap(\.fileExtensions))
+    }
+
+    /// Look up the language for a given file extension
+    ///
+    /// - Parameter ext: A file extension (without the leading dot), e.g. `"swift"`, `"java"`, `"cpp"`
+    /// - Returns: The matching language, or `nil` if the extension is not supported
+    public static func language(forFileExtension ext: String) -> SnippetLanguage? {
+        let lowered = ext.lowercased()
+        return allCases.first { $0.fileExtensions.contains(lowered) }
+    }
+}
--- a/Sources/Snippets/Model/SnippetLanguage.swift
+++ b/Sources/Snippets/Model/SnippetLanguage.swift
@@ -27,6 +27,11 @@ public enum CommentStyle: Sendable {
     public static var xml: Self {
         .blockComment(prefix: "<!--", suffix: "-->")
     }
+
+    /// C-style block comment `/* ... */`
+    public static var cBlock: Self {
+        .blockComment(prefix: "/*", suffix: "*/")
+    }
 }
 
 /// A programming language supported for code snippets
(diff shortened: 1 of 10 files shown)


ktoso:
@swift-ci please smoke test

Sources/Snippets/Model/SnippetLanguage.swift
```diff
@@ -0,0 +1,80 @@
...
+        case .java:       return "java"
+        case .kotlin:     return "kotlin"
+        case .c:          return "c"
+        case .cpp:        return "c++"
+        case .objectiveC: return "objective-c"
+        case .csharp:     return "csharp"
+        case .go:         return "go"
+        case .rust:       return "rust"
+        case .javascript: return "javascript"
+        case .typescript:  return "typescript"
```
ktoso:
```suggestion
        case .typescript: return "typescript"
```

Tests/SwiftDocCPluginUtilitiesTests/SnippetParseTests.swift
```diff
@@ -461,45 +461,237 @@ class SnippetParseTests: XCTestCase {
...
+        #include <iostream>
+        // snippet.show
+
+        int main() {
+            std::cout << "Hello" << std::endl;
+            return 0;
+        }
+        """
+
+        let snippet = Snippet(parsing: source, sourceFile: SnippetParseTests.fakeSourceFilename)
```
heckj: completely a nit - but I think `SnippetParseTests.fakeSourceFilename` resolves to "Test.swift" in this case - wouldn't we want to that to be something ending in .cpp?

heckj (commented):
Love it - and... the questions:

This kicks the door open for any file type, with a set of common ones that we may well run into (or have some expectations that we might), but there's always the fallback there. We started with "Swift and only swift", and this jumps to "any file, mapping comment formats to expected patterns - but if we're not sure or have never seen it before, fall back to the good ole `//` pattern and hope for the best.

Do we want take this as an "anything, anywhere" as an extension (for example, what about a markdown file or JSON file in the Snippets directory - as kind of crazy meta as that is, right now I think that means it'd get converted into a symbol graph based on it's name too)? The alternative that I see if a more limited expansion - pinning down the specific of files we do want to support, and either excluding (or possibly warning) on extraneous files from being processed. The pathological case is someone goes a bit crazy with what's in the Snippets directory, and now each build is iterating through the whole of that directory and processing *all* of those files into symbolgraphs for possible inclusion into catalogs.

My concern is that this goes too wide, too quickly - and I'm concerned that the unexpected/unintentional side effects for accidentally including non-code/text files in Snippets, thinking that would just magically work, and how to handle that. I'm not convinced I'm correct or not being overly paranoid on this front, but I wonder if we shouldn't be more conservative with what additional modes we enable, even if that's a large set - and the be more specific about "ignore anything else"

Related: there is a `SnippetLanguage.supportedFileExtensions` that appears to be unused, but may be good to this end at SymbolGraph+Snippet.swift:21-22 where we attempt to recognize the type of the file from it's extension. Likewise, if we want to constrain the supported sets, it seems like it would be a good filter for `_findSnippetFilesInDirectory` to use)

Based on the tests, it looks like the code is set up to allow parsing of comments from just about any form - so long as it matches the snippet pattern we expect to see - so you can jump back and forth between `//` and `/* */` or other comment formats if the file/language identified allows it - and use any of those for the snippet "parsing" for controls: hide, show, and slice name.

heckj:
(I continued poking after my initial review, and I think there may be some lingering issues in the possibility of enabling multi-line comments and parsing from them - the original design expected single-line comments, and did some wrangling of indent based on those. This really bites when applying "explanation lines" to snippet code, but something I want to dig into a bit more - which touches on this PR

ktoso:
The multiline comments really still are single line but I accept the /* format… so it’s pretty naive but imho good enough for the snippet marking.

yeah I do think we should allow “anything anywhere”.

this has been a problem ever since the first proposal I’m requesting to address: I want to use pieces of my tests in snippets. Snippet code isn’t validated the same way as tests are and we need this imho.

Tests/SwiftDocCPluginUtilitiesTests/SnippetParseTests.swift
ktoso: Oh let me check that, thanks

Sources/Snippets/Model/SnippetLanguage.swift
```diff
@@ -0,0 +1,139 @@
...
+    case scala
+    case groovy
+    case python
+    case bash
+    case zsh
+    case xml
+    case html
+
+    /// The language identifier used in symbol graphs and for syntax highlighting
+    public var id: String {
```
d-ronnqvist, line 67: minor: with the exception of `.cpp` and `.objectiveC`, the `id` is always the same as the `rawValue` derived from spelling of the case. Would it make sense to customize the raw values for those two cases:
```swift
case cpp = "c++"
case objectiveC = "objective-c"
```
and make the `id` property return the `rawValue`:
```swift
public var id: String {
    rawValue
}
```
instead of repeating all the identifier names in this switch statement?

Sources/Snippets/Model/SnippetLanguage.swift
```diff
@@ -0,0 +1,139 @@
+// This source file is part of the Swift.org open source project
+//
+// Copyright (c) 2024 Apple Inc. and the Swift project authors
```
d-ronnqvist: nit: This is likely meant to be 2026.

Sources/SwiftDocCPluginUtilities/Snippets/SnippetExtractor.swift
```diff
@@ -78,7 +78,7 @@ public class SnippetExtractor {
         }
         var snippetInputFiles = [String]()
         for case let potentialSnippetURL as URL in snippetEnumerator {
-            guard potentialSnippetURL.pathExtension.lowercased() == "swift" else {
+            guard !potentialSnippetURL.pathExtension.isEmpty else {
```
d-ronnqvist, line 81: question: Would it makes sense to check that the file extension is in `supportedFileExtensions` and continuing (skipping the file) otherwise?

Sources/Snippets/Model/SnippetLanguage.swift
```diff
@@ -0,0 +1,139 @@
...
+// Licensed under Apache License v2.0 with Runtime Library Exception
+//
+// See https://swift.org/LICENSE.txt for license information
+// See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+
+/// The comment syntax used by a language for snippet markers and explanations
+public enum CommentStyle: Sendable {
+    /// Line comment with a given prefix, e.g. `//` or `#`
+    case lineComment(String)
+    /// Block comment with a prefix and suffix, e.g. `<!--` and `-->`
```
d-ronnqvist, line 13: question: Because the implementation (currently?) requires that the prefix and suffix are on the same line, where do you feel that it would make sense to document that restriction (both in the implementation and in the user-facing documentation about snippets)?

Sources/Snippets/Parsing/SnippetParser.swift
```diff
@@ -214,17 +216,71 @@ extension SnippetParser {
...
+        switch commentStyle {
+        case .lineComment(let prefix):
+            guard trimmed.trimExpectedPrefix(prefix) else { return false }
+        case .blockComment(let prefix, let suffix):
+            guard trimmed.trimExpectedPrefix(prefix) else { return false }
+            if let closingRange = trimmed.range(of: suffix, options: .backwards) {
+                trimmed = trimmed[..<closingRange.lowerBound]
+                while trimmed.last?.isWhitespace == true {
+                    trimmed = trimmed.dropLast()
+                }
```
d-ronnqvist, line 360: minor: This is likely going to do the right thing in practice but with a block it's not guaranteed that there's nothing else after the comment suffix. For example: ` /* snippet.hide */ let pi = 3.14 `. 

I don't know if that's a possibility that we care about supporting (considering that the only issue that I can see with the implementation is that it would trim the "wrong" whitespace) but it might be a worth explicitly deciding that we don't care about that edge case and add a code comment that acknowledges that trimming the whitespace at the end _could_ do remove the "wrong" whitespace in that situation. That makes it easier for someone else who looks at the code later to understand what the implementation is doing and why.

d-ronnqvist (commented):
This generally looks good to me. Just a few minor comments. 

I also wonder where (if a all) it would make sense to test that the snippet extractor excludes files without any file extension and/or files with an unsupported file extension (like `.txt` or `.png`).

ktoso:
Thank you for reviewing! I'll follow up on it

Sources/Snippets/Parsing/SnippetParser.swift
ktoso: Maybe we could warn if we see this happen? I'll see what I can come up with

Sources/SwiftDocCPluginUtilities/Snippets/SnippetExtractor.swift
ktoso: Hmm... I worry about this actually, what if we want to support `.pkl` snippets or something else the tool does not know about 🤔 I'll think what'll happen, and this also reminded me to add `yaml` to supported languages as well.

Sources/Snippets/Model/SnippetLanguage.swift
ktoso: Thanks!

Sources/Snippets/Model/SnippetLanguage.swift
ktoso: In implementation probably here on the case of the enum... 

In user facing docs we probably need to do that part of the work in docc documentation where Snippets are documented 🤔 I could follow up on it

Sources/Snippets/Parsing/SnippetParser.swift
ktoso: I added warnings about this situation, it'll be like: 

```
  Example.html:3: warning: content after block comment snippet marker will be ignored: 'let code here = 12'
```

Sources/Snippets/Model/SnippetLanguage.swift
ktoso: Uff, I reworked this to allow things like

```
/*

snippet.show

*/
```

etc. So this will now work if the comment contains ONLY the marker command (`snippet.[...]`). [`7c6ead4` (this PR)](https://github.com/swiftlang/swift-docc-plugin/pull/120/commits/7c6ead40fccb50726cd2ba423992d2531325d3fd)

Sources/snippet-extract/SnippetBuildCommand.swift
```diff
@@ -145,6 +150,14 @@ struct SnippetExtractCommand {
             throw error
         }
     }
+
+    func printWarning(_ message: String) {
+        guard let msg = "\(message)\n".data(using: .utf8) else {
+            return 
+        }
+        
+        FileHandle.standardError.write(msg)
```
d-ronnqvist: FYI (non blocking): It is always possible to convert the UTF8 representation of a string to data but the `data(using:)` method cannot reflect that (it has to have an optional return type to account for other encodings). Instead you can pass the UTF8 view into the `Data` initializer and you won't have to deal with an optional result:

```suggestion
        FileHandle.standardError.write(Data(message.utf8))
```

Sources/Snippets/Parsing/SnippetParser.swift
```diff
@@ -214,19 +338,187 @@ extension SnippetParser {
...
+            if !afterSuffix.allSatisfy(\.isWhitespace) {
+                return afterSuffix.drop { $0.isWhitespace }
+            }
+        }
+        return nil
+    }
+}
+
+/// A warning emitted during snippet parsing
+public struct SnippetWarning: Sendable {
```
d-ronnqvist, line 511: Considering that this target isn't a product, meaning that even if it's a public type there's nothing outside this package breaks if we modify or even remove this type in the future I think this warning design is fine. It does what it needs to quite minimally and doesn't need to try to accommodate future changes.

I would possibly suggest renaming the `warning` and `text` properties because those threw me off for a second while reading `testWarningOnTrailingContentAfterBlockCommentMarker()`. That said, I don't consider that blocking feedback because we can always rename them in the future.

d-ronnqvist (approved):
Thanks for working on this enhancement. This looks good to me.

d-ronnqvist:
@swift-ci please test

Sources/snippet-extract/SnippetBuildCommand.swift
ktoso: Oh, thanks! I always used that other route, TIL

Sources/Snippets/Parsing/SnippetParser.swift
ktoso: Let me remove the publics, sorry about that, no reason to have them.

Hmmm the names I kinda liked, any preferences what to call it?

ktoso:
Thanks for reviewing I'll follow up shortly with the small cleanups

Sources/Snippets/Parsing/SnippetParser.swift
d-ronnqvist: I'm not sure. Going by the documentation comments of the properties I would have possibly suggested `description` and `sourceText` but it makes sense for the `description` to be the representation that's formatted to be printed on the command line. Maybe `message` could be a possible name for the warning's short description text?

ktoso:
@swift-ci please test

ktoso:
@swift-ci please test

ktoso:
@swift-ci please test

ktoso:
@swift-ci please test

ktoso:
@d-ronnqvist could you help me out there? I can't reproduce the CI issues `./bin/test` is passing, these failures seem unrelated...?

d-ronnqvist:
I opened [this draft PR with no changes compared to main](https://github.com/swiftlang/swift-docc-plugin/pull/121) to see if that reproduces these CI failures.

ktoso:
Good idea, let's see; Yeap, same failures AFAICS

ktoso:
Trying my luck if main got unbroken

ktoso:
@swift-ci please test -- nope, main still broken

d-ronnqvist:
@swift-ci please test

d-ronnqvist:
This CI is unblocked after applying [this workaround](https://github.com/swiftlang/swift-docc-plugin/pull/122/changes) (in another, already merged, PR).

ktoso:
Thank you!

PR #126, opened 2026-05-24
Ignore snippet markers inside multiline string literals

adityasingh2400:
Bug/issue #: swiftlang/swift-docc#946

## Summary

Snippet slice markers are line comments of the form `// snippet.show`, `// snippet.hide`, `// snippet.<name>`, and `// snippet.end`. The parser matched them line by line with no awareness of string literals, so a line like `// snippet.hide` that appears inside a Swift multiline string literal was treated as a real marker. That cut the literal at the marker line and dropped the surrounding text, and in some arrangements it left unbalanced `"""` delimiters in the presented code.

For example, this snippet:

```swift
// snippet.show
let text = """
Only this part of the code will be shown.

// snippet.hide

Snippet markers inside multiline string literals are ignored.

// snippet.show
"""
```

previously rendered with the literal sliced at the inner `// snippet.hide` / `// snippet.show` lines instead of showing the whole string.

This adds a small per-language string literal scanner. It is fed each source line in order and reports whether the start of the next line lies inside an open multiline string literal, including raw literals such as `#"""..."""#` where the closing delimiter must carry the matching number of `#` characters. While the parser is inside such a literal it emits the line as presentation content and never interprets it as a marker or comment.

The string literal syntax is language specific (`StringLiteralSyntax`) and is currently enabled for Swift only, looked up by file extension alongside the existing comment style lookup. Languages without a known multiline string syntax are unaffected, so the `#`, XML, and other comment styles behave exactly as before.

The scanner intentionally does not track delimiters that appear inside block comments, which is not a realistic case for snippet markers and is noted in the code.

## Dependencies

None. The fix and tests are contained in this repository.

## Testing

Steps:
1. `swift test --filter SnippetParse`

New tests in `Tests/SwiftDocCPluginUtilitiesTests/SnippetParseTests.swift` cover:
- markers inside a plain multiline string literal are ignored and the literal is shown intact
- markers resume working on lines after a multiline literal closes
- raw multiline literals (`#"""..."""#`), where a bare `"""` does not close the literal
- a single-line string containing marker-like text on a code line is left untouched
- a non-Swift language (shell, `#` comments) does not treat `"""` as a literal delimiter
- a slice whose body contains a multiline literal keeps a correct line range

The full `SwiftDocCPluginUtilitiesTests` suite passes (83 tests). I ran `swift test` with `SWIFT_SKIP_BUILDING_UPSTREAM_DOCC=true`; I did not run the full `./bin/test` integration path that builds upstream swift-docc, since this change is exercised entirely by the unit tests.

## Checklist

- [x] Added tests
- [ ] Ran the `./bin/test` script and it succeeded (ran `swift test` for the affected target; did not run the upstream-DocC integration path)
- [x] Updated documentation if necessary (no user-facing docs change; the behavior now matches the documented marker semantics)

### DIFF
--- a/Sources/Snippets/Model/Snippet.swift
+++ b/Sources/Snippets/Model/Snippet.swift
@@ -34,7 +34,13 @@ public struct Snippet {
 
     init(parsing source: String, sourceFile: URL) {
         let commentStyles = SnippetLanguage.commentStyles(forFileExtension: sourceFile.pathExtension)
-        let extractor = SnippetParser(source: source, commentStyles: commentStyles, sourceFile: sourceFile.lastPathComponent)
+        let stringLiteralSyntax = SnippetLanguage.stringLiteralSyntax(forFileExtension: sourceFile.pathExtension)
+        let extractor = SnippetParser(
+            source: source,
+            commentStyles: commentStyles,
+            stringLiteralSyntax: stringLiteralSyntax,
+            sourceFile: sourceFile.lastPathComponent
+        )
         self.explanation = extractor.explanationLines.joined(separator: "\n")
         self.presentationLines = extractor.presentationLines.map(String.init)
         self.slices = extractor.slices
--- a/Sources/Snippets/Model/SnippetLanguage.swift
+++ b/Sources/Snippets/Model/SnippetLanguage.swift
@@ -39,6 +39,35 @@ public enum CommentStyle: Sendable {
     }
 }
 
+/// The multiline string literal syntax used by a language.
+///
+/// Snippet markers are line comments such as `// snippet.show`. When such a
+/// sequence appears inside a multiline string literal it is literal text, not a
+/// marker, so the parser must know how a language opens and closes these
+/// literals in order to ignore markers found within them.
+public struct StringLiteralSyntax: Sendable {
+    /// The delimiter that opens and closes a multiline string literal, e.g. `"""`.
+    public var multilineDelimiter: String
+
+    /// Whether the language allows raw-string customization of the delimiter by
+    /// surrounding it with one or more `#` characters, e.g. Swift's `#"""..."""#`.
+    ///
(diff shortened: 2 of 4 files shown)


d-ronnqvist:
It would be good to discuss this change in behavior with a broader audience in the Swift Forums before reviewing this code.

I'm not sure where I stand on the topic if the snippet parser _should_ be syntactically aware or not. 
It would help to know if there's a use case for adding a `// snippet.hide` marker inside of a multiline string literal.

adityasingh2400:
Thanks for taking a look, and fair question on the use case. The motivation isn't someone deliberately writing `// snippet.hide` inside a string. It is that a multiline string literal can contain those lines as ordinary content, and the parser currently treats them as real markers. That is the bug filed in swiftlang/swift-docc#946.

The most natural case is a snippet that documents the snippet markers themselves (the example in the description), but it generalizes to any multiline string whose lines happen to match the marker pattern. Today that silently slices the literal at the inner marker lines and drops text, and in some arrangements leaves unbalanced `"""` in the presented code, so the rendered snippet is wrong even though nothing in real code was marked.

On whether the parser should be syntactically aware: this stays deliberately minimal. It tracks only multiline string delimiters (`"""`, plus Swift's raw `#"""..."""#`), not full syntax, and it only ignores markers found inside such a literal, where they can never be a real directive. Intended marker usage is unaffected.

Happy to start a Swift Forums thread with this repro to get broader input before you review, if you would still like that. Just say the word.

d-ronnqvist:
Yes, I think it would be good to start a Swift Forums thread to get more inputs.

adityasingh2400:
Sounds good. I will start a thread on the Swift Forums under Development > Swift-DocC, linking this PR and the underlying issue swiftlang/swift-docc#946, and post the link back here so others can weigh in on whether the parser should stay minimal or become more syntactically aware.

